PR #8, opened 2021-02-03
DAPI-23 Add swagger docs and change licence to MIT

electrified:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsdeliusapi/config/SwaggerConfig.kt
```diff
@@ -0,0 +1,49 @@
...
+              .or(PathSelectors.regex("(\\/V1.*)"))
+          )
+      )
+      .build()
+
+  private fun contactInfo(): Contact =
+    Contact(
+      "HMPPS Digital Studio",
+      "",
+      "dps-hmpps@digital.justice.gov.uk"
```
mxco86, line 35: Do DPS want all our support emails :) I guess we need a team mailbox

mxco86 (approved):
👍 plus a bonus JIRA ticket

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsdeliusapi/config/SwaggerConfig.kt
electrified: I had a look if there was a suitable group - but couldn't find one, so set it the same as nDelius

PR #21, opened 2021-02-11
DAPI 14 validation

axle-h:


src/test/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/ContactServiceTest.kt
```diff
@@ -25,30 +35,111 @@ class ContactServiceTest {
...
+  }
+
+  @Test
+  fun `Attempting to create contact with type`() {
+    havingDependentEntities(havingType = false)
+    shouldThrowBadRequest()
+  }
+
+  @Test
+  fun `Attempting to create contact with outcome`() {
```
electrified, line 89: Are some  of these test named wrongly, e.g. should be "without outcome"?

docker-compose.yml
```diff
@@ -16,15 +16,23 @@ services:
...
     environment:
       - SERVER_PORT=8080
       - SPRING_PROFILES_ACTIVE=dev
 
+  oracle:
+    image: 895523100917.dkr.ecr.eu-west-2.amazonaws.com/hmpps/delius-test-db:latest
+    networks:
+      - hmpps
+    ports:
+      - "1521:1521"
```
marcus-bcl, line 35: What happens for people without ECR access? 

It looks like this starts up the oracle db but doesn't use it by defauly (`SPRING_PROFILES_ACTIVE=dev`). Might be worth using a docker-compose [profile](https://docs.docker.com/compose/profiles/) to only start up and use the oracle container conditionally?

electrified (approved):

electrified:
Looks good!

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/ContactService.kt
```diff
@@ -4,39 +4,57 @@ import org.springframework.stereotype.Service
...
   fun createContact(request: NewContact): ContactDto {
-    val offender = offenderRepository.findByCrn(request.offenderCrn).orElse(null)
-      ?: throw BadRequestException("Offender with crn ${request.offenderCrn} does not exist")
+    val type = contactTypeRepository.findByCode(request.contactType)
+      ?: throw BadRequestException("Contact type with code '${request.contactType}' does not exist")
 
-    val contact = Contact(
-      offender = offender,
+    if (request.alert && !type.contactAlertFlag) {
+      throw BadRequestException("Contact type '${type.code}' is not support alert")
```
marcus-bcl: Do we need to standardise error messages somewhere, rather than hardcoding strings? Probably don't need to worry about it right now but could be something to think about. Would make it easier to cross-reference with the Delius properties files.

marcus-bcl (approved):
Looks really good. Approved with a couple of comments.

docker-compose.yml
axle-h: fixed, see latest commit. this is now opt-in.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/ContactService.kt
axle-h: leaving this for later. I have a list of codes that we should probably support but error messages may be out of scope.

PR #26, opened 2021-02-11
DAPI-25 auditing

electrified:


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/audit/AuditService.kt
```diff
@@ -0,0 +1,81 @@
...
+      mapOf(AuditParameter.OFFENDER_ID to offenderId.toString()),
+      true
+    )
+
+  fun failedInteraction(userId: Long, offenderId: Long, interaction: AuditableInteraction) =
+    createAuditedInteraction(
+      LocalDateTime.now(),
+      userId,
+      interaction,
+      mapOf(AuditParameter.OFFENDER_ID to offenderId.toString()),
```
marcus-bcl, line 30: Just a note, when it comes to doing NSIs we'll have to pass `mapOf(AuditParameter.NSI_ID to entity.id.toString())` here rather than offenderId. Does it make sense to move the creation of the map up to the caller, so we can pass any map of audit params depending on the context?

Doesn't need to be done now, but something to be aware of as we add more stuff.

marcus-bcl (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/audit/AuditService.kt
```diff
@@ -0,0 +1,81 @@
...
+    return builder.toString()
+  }
+
+  private fun isInteractionAuditable(enabledDate: LocalDateTime?) =
+    enabledDate != null && LocalDateTime.now().isAfter(enabledDate)
+
+  companion object {
+    private const val INTERACTION_SUCCEEDED = "P"
+    private const val INTERACTION_FAILED = "F"
+  }
```
axle-h, line 80: lol, is this yet another boolean storage type? Hibernate has the yes_no type, maybe we can add our own pass_fail type to avoid this nonsense?!

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/audit/AuditService.kt
axle-h: I would provide offenderId and NSI id as optional params, throw if at least one is not populated and build the map in the function. Otherwise youre leaking the keys everywhere, making testing difficult and losing control of what's actually going in this map.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/audit/AuditableInteraction.kt
```diff
@@ -0,0 +1,5 @@
+package uk.gov.justice.digital.hmpps.deliusapi.service.audit
+
+enum class AuditableInteraction(val code: String) {
+  ADD_CONTACT("CLBI003"),
+}
```
axle-h, line 5: ouch. no way of retrieving this code from the db at app start up?

axle-h (approved):
looks fine to me, just a minor kotlin-ify
some things that can be improved with significant effort...

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/audit/AuditService.kt
electrified: It is indeed :D Will look at that

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/audit/AuditableInteraction.kt
electrified: I think the code e.g. `CLBI003` is the minimal amount of data we can get away with in code. From there you can go and get the other details from the BUSINESS_INTERACTION table

PR #29, opened 2021-02-12
fix: check start date where also checking end date

axle-h:


marcus-bcl (approved):
Do we need tests for these conditions? Might be useful for if we switch to using `@Filter`.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/entity/OfficeLocation.kt
```diff
@@ -11,7 +11,7 @@ import javax.persistence.Table
 
 @Entity
 @Table(name = "OFFICE_LOCATION")
-@Where(clause = "END_DATE IS NULL OR END_DATE > CURRENT_DATE")
+@Where(clause = "START_DATE <= CURRENT_DATE AND (END_DATE IS NULL OR END_DATE > CURRENT_DATE)")
```
electrified, line 14: I think these checks are fine for new appointments, but historic appointments may be linked to office locations that are no longer current.

Do we not care as we don't have any get endpoints?

electrified: I think the same consideration applies to staff and team.

electrified: Or does the where not get applied once there is an associated entity linked by ID?

axle-h: Hmm my understanding of this was that start & end dates act as soft deletion.
Delius checks them at least for team. I didnt confirm with office location. Ill have a look now.

axle-h:
> Do we need tests for these conditions? Might be useful for if we switch to using `@Filter`.

yeah... we really need better seed data. Maybe I can search the existing seed for appropriate locations, teams & staff...

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/entity/OfficeLocation.kt
marcus-bcl: Good point, in the Delius code it doesn't do the start date check for office location (from what I can tell).

PR #32, opened 2021-02-15
DAPI-18 Validation: Contact Type to Contact Outcome link

marcus-bcl:


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/ContactService.kt
```diff
@@ -28,6 +28,10 @@ class ContactService(
     val type = contactTypeRepository.findByCode(request.contactType)
       ?: throw BadRequestException("Contact type with code '${request.contactType}' does not exist")
 
+    if (type.outcomeTypes?.none { it.code == request.contactOutcome } == true) {
```
electrified, line 31: `R_CONTACT_OUTCOME_TYPE` has a `selectable` field. Should we filter on that too?

electrified (approved):

axle-h (approved):

PR #35, opened 2021-02-16
DAPI-44: Reference data to support insert Contact endpoint

MichaelWetherallBCL:


marcus-bcl:
Is it correct that we don't have an outcome linked to this type? If so, restricting to this type now might make it difficult to test the outcome stuff we've done... Same for attendance/location etc. @axle-h any thoughts?

marcus-bcl:
Also - is it worth giving the file a more specific name eg. `interventions_contact_reference_data` ?

I'll have a look at which (if any) of the other data sql files we need to keep. Looks like they contain test data specific to the community-api tests that we probably don't need.

mxco86 (approved):
👍 Matches what has been discussed

PR #36, opened 2021-02-17
refactor: run auditing via aoc

axle-h:


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/audit/AuditContext.kt
```diff
@@ -0,0 +1,15 @@
+package uk.gov.justice.digital.hmpps.deliusapi.service.audit
+
+import kotlin.concurrent.getOrSet
+
+data class AuditContext(
+  var offenderId: Long? = null,
+  var nsiId: Long? = null,
```
marcus-bcl, line 7: Is this going to limit us in the future? Each auditable interaction has a potentially different context which can be made up of any number of IDs. If all we're doing is combining them into a string for informational purposes, would it be better to leave it up to the caller to decide what fields go into the context?

axle-h: the caller does decide, see AuditService::25

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/advice/AuditableAspect.kt
```diff
@@ -0,0 +1,25 @@
...
+annotation class Auditable(val interaction: AuditableInteraction)
+
+@Aspect
+@Component
+class AuditableAspect(private val auditService: AuditService) {
+  @AfterReturning("@annotation(auditable)")
+  fun auditSuccess(auditable: Auditable) =
+    auditService.successfulInteraction(auditable.interaction, AuditContext.get())
+
+  @AfterThrowing("@annotation(auditable)")
```
marcus-bcl, line 22: Much prefer this pattern 👍

Does this handle an exception on transaction commit? Are we forcing ourselves to rely on saveAndFlush here? Guess we can deal with that when we properly implement transactions...

axle-h: yeah Im not sure how the transaction aop will affect this, I presume ordering will have to be configured somehow. Something to solve with transactions.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/audit/AuditContext.kt
axle-h: just noticed that once we start nesting audited interactions this thread local storage will create issues. Another thing we can solve later.. or just rip out for something function local.

marcus-bcl: But they only get to choose offenderId or nsiId don't they? There are hundreds of other interactions that have their own audit context made up of different fields than those two - so might be more future-proof to allow the caller to pass a Map of Strings?

marcus-bcl: I don't think there will be a situation where we need to nest interactions, however we may just need to keep a map of InteractionCode -> AuditContext in the ThreadLocal.

axle-h: nothing stopping us adding more optional properties to this context - it's basically a strongly typed map at this point.
I much prefer this to magic strings.
But I have no context on how big this context will grow. Do you really think that we will implement hundreds of interactions in this api and across the subset of interactions that we implement, how many unique keys are there really?

marcus-bcl: Happy to be pushed back on it, if this way works then let's stick with it 😄 ...just worries me slightly whenever we deviate from what Delius does!

axle-h: yeah.. we can refactor this away if it becomes an issue (like we end up with 100's of properties), which will be easy - due to the issues around testing aspectj we aren't actually testing the property setters instead just that the context is passed successfully to the audit service.

I just pushed a fix for nested audited interactions

axle-h: we will need to nest interactions e.g. creating an nsi also creates contacts... although we may be able to pull this up into a wrapper service. I think it's still worth having support present just in case.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/advice/AuditableAspect.kt
electrified: What happens if the exception is thrown before the context is set?

axle-h: audit failure record will not be associated to anything, should probably set asap within each function, Ill move the offender one up closer to where we retrieve the offender

axle-h: it's not an issue here (the offender check can go first) but in some case we may want to aggregate exceptions up to avoid not writing the context

axle-h: pushed a fix but now I realise that the audit service will actually throw if there are no keys in the parameter map. So we need to make sure we dont even cal the service if no context actions have been performed by the throwing method.

electrified: Yes, it feels like it's an unlikely but potential scenario

e.g. The Offender table is corrupted/locked so we can't fetch an offender - resulting in an exception prior to us populating the parameters. I guess logging is the best we can do in this scenario.

electrified (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/advice/AuditableAspect.kt
axle-h: even more likely scenario -> the offender does not exist!
Ive just pushed a fix & test for this

PR #41, opened 2021-02-18
DAPI-77 - Make start and end time non-mandatory and dependant on cont…

electrified:
…act type and outcome presence


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/dto/v1/ContactDto.kt
```diff
@@ -9,14 +9,14 @@ data class ContactDto(
   val id: Long,
   val offenderCrn: String,
   val type: String,
-  val outcome: String,
+  val outcome: String?,
```
electrified, line 12: This was missed from my previous PR. Have changed tests to catch this kind of error.

axle-h (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/ContactService.kt
```diff
@@ -90,6 +90,14 @@ class ContactService(
     val staff = team.staff?.find { it.code == request.staff }
       ?: throw BadRequestException("Staff with officer code '${request.staff}' does not exist in team '${request.team}'")
 
+    if (type.attendanceContact && request.startTime == null) {
+      throw BadRequestException("Contact type  '${type.code}' requires a start time")
+    }
+
+    if (type.recordedHoursCredited && outcome != null && request.endTime == null) {
+      throw BadRequestException("Contact type '${type.code}' requires an time when an outcome is provided")
```
marcus-bcl, line 98:
```suggestion
      throw BadRequestException("Contact type '${type.code}' requires a time when an outcome is provided")
```

marcus-bcl (approved):

PR #48, opened 2021-02-22
feat: skeleton create nsi endpoint

axle-h:


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/validation/RelativeDate.kt
```diff
@@ -0,0 +1,83 @@
...
+
+enum class RelativeDateType {
+  PAST,
+  FUTURE,
+}
+
+@Target(AnnotationTarget.FIELD)
+@Constraint(validatedBy = [RelativeDateValidator::class])
+@Retention(AnnotationRetention.RUNTIME)
+annotation class RelativeDate(
```
marcus-bcl, line 19: I like the use of custom validators, it's a nice way to capture the repeated logic. Is it possible to use the built-in `@Past` and `@Future` validators for these ones?

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/validation/StaffCode.kt
```diff
@@ -0,0 +1,19 @@
...
+import kotlin.reflect.KClass
+
+@Target(AnnotationTarget.FIELD)
+@Retention(AnnotationRetention.RUNTIME)
+@Constraint(validatedBy = [])
+@ReportAsSingleViolation
+@NotBlankWhenProvided
+@Size(min = 7, max = 7)
+annotation class StaffCode(
+  val message: String = "must be a valid staff code",
```
marcus-bcl, line 16: Staff code should also be alphanumeric. Might be easier to use a `@Pattern` validator for this, similar to the user management tool (https://github.com/ministryofjustice/ndelius-um/blob/20980240b7085e815149f9577ac33877d3bd5cfe/src/main/java/uk/co/bconline/ndelius/model/User.java#L68-L69)

marcus-bcl: 🤔 probably safe to assume the other codes should be alphanumeric too, we should check though. I guess we're not actually creating anything with these values, just checking they exist, so it's not that important really.

src/main/resources/db/data/V1_20__nsi_business_interaction_data.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO BUSINESS_INTERACTION (BUSINESS_INTERACTION_ID, BUSINESS_INTERACTION_CODE, DESCRIPTION, ENABLED_DATE, AUDIT_INTERACTION_PARAMETER, ROW_VERSION)
```
marcus-bcl, line 1: We should add this to [https://github.com/ministryofjustice/hmpps-delius-api/blob/main/src/main/resources/db/data/V1__reference_data/V1_3__business_interactions.sql](https://github.com/ministryofjustice/hmpps-delius-api/blob/main/src/main/resources/db/data/V1__reference_data/V1_3__business_interactions.sql) instead. I know it kind of breaks the Flyway versioned migration model, but given it's only a dev database I think having organised files that can be made sense of is more important.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/dto/v1/NewNsi.kt
```diff
@@ -0,0 +1,180 @@
...
+import javax.validation.constraints.Size
+
+/**
+ * A request to create a new non-statutory intervention.
+ */
+@TimeRanges
+@FieldGroups
+data class NewNsi(
+  /**
+   * The type of the NSI.
```
marcus-bcl, line 28: Nice, should/could we add these comments to the Swagger docs? I think converting these comments to `@ApiModelProperty` annotations might do it.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/dto/v1/NewNsi.kt
```diff
@@ -0,0 +1,180 @@
...
+   * TODO must be a valid sub type of the specified type
+   * TODO required when specified nsi type has >0 selectable sub types
+   */
+  @field:NsiTypeCode
+  val subType: String? = null,
+
+  /**
+   * The offender CRN.
+   *
+   * TODO existing, non-soft deleted offender
```
marcus-bcl, line 52: You've probably already thought of this, but once we mark the controller methods as `@Transactional` we should be able to use validators for these validations that check the database. This is because any repeated DB calls would get the result from the transaction/session cache, so there would be no extra DB hit.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/dto/v1/NewNsiManager.kt
```diff
@@ -0,0 +1,35 @@
...
+   */
+  @field:StaffCode
+  val staff: String,
+
+  /**
+   * Team managing this intervention
+   *
+   * TODO must be existing, selectable (via start & end date) team
+   * TODO must be in specified managerProvider
+   * TODO cant we derive this from the staff?!
```
marcus-bcl, line 22: I'm not sure we can... staff members can be in multiple teams

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/dto/v1/NewNsiManager.kt
```diff
@@ -0,0 +1,35 @@
...
+   */
+  @field:TeamCode
+  val team: String,
+
+  /**
+   * Provider managing this intervention
+   *
+   * TODO must be existing, selectable (via selectable flag) provider
+   * TODO cant we derive this from the team?!
```
marcus-bcl, line 31: I feel like we should be able to, but Delius doesn't and CRCs also must pass the provider along with staff and team, so maybe there is a reason for it. Could it be something like Team codes aren't unique across providers?

src/test/kotlin/uk/gov/justice/digital/hmpps/deliusapi/validation/NewContactValidationTest.kt
```diff
@@ -1,124 +1,48 @@
...
+    @JvmStatic
+    private fun validCases() = ValidationTestCaseBuilder.from(NewContact::class, valid = true)
+      .kitchenSink()
+      .string(NewContact::outcome) { it.isNull().length(1).length(10) }
+      .string(NewContact::officeLocation) { it.isNull() }
+      .time(NewContact::endTime) { it.bothNull(NewContact::startTime).isNull() }
+      .string(NewContact::notes) { it.isNull().empty() }
+      .string(NewContact::description) { it.isNull().empty() }
+      .number(NewContact::requirementId) { it.isNull().bothNull(NewContact::eventId) }
+      .cases
```
marcus-bcl, line 30: Cool 👍

src/test/resources/application-test.yml
```diff
@@ -6,4 +6,7 @@ management.endpoint:
   info.cache.time-to-live: 0
 
 tokenverification:
-  api.base.url: http://localhost:9090/auth
\ No newline at end of file
+  api.base.url: http://localhost:9090/auth
+
+contacts:
+  allowed-types: C001,C002,C003,COUP
```
marcus-bcl, line 12: Do we need these here, if we're specifying them in the `@TestPropertySource` on the test? I tried to avoid restricting the types for all tests, only the ones where I needed to check the validation.

marcus-bcl (dismissed):

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/validation/RelativeDate.kt
axle-h: ah, didn't spot those, will take a look

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/validation/StaffCode.kt
axle-h: nice, yeah this was missing from the original contact ticket, will add

src/main/resources/db/data/V1_20__nsi_business_interaction_data.sql
axle-h: looks like this was missed when rebasing onto your data cleanup work

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/dto/v1/NewNsi.kt
axle-h: yeah, it's not a bad shout and would improve the reporting of errors as they would be aggregated by default... however, it would make testing more difficult. I also like the current split between field level validation and business logic level validation we have right now. There could be an idiomatic way of keeping this split with everything in validation annotations though. One to revisit later.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/dto/v1/NewNsiManager.kt
axle-h: ah ok will remove

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/dto/v1/NewNsiManager.kt
axle-h: ok will remove

src/test/kotlin/uk/gov/justice/digital/hmpps/deliusapi/validation/NewContactValidationTest.kt
axle-h: yeah kotlin reflection made this quite easy to do efficiently (it's one round of reflection for all test cases)

src/test/resources/application-test.yml
axle-h: it's for the integration tests. I can remove it here and add to the test directly?

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/dto/v1/NewNsi.kt
marcus-bcl: True - maybe that's a sensible place to draw the line for now. Otherwise we might end up over-abstracting stuff to fit into validators.

src/main/resources/db/data/V1_20__nsi_business_interaction_data.sql
axle-h: actually I purposely kept it separate from that existing migration so that you guys didnt have a wtf moment when checking this out and the integration tests fail locally. Since you asked though I will merge it into the existing migration, just remember you'll need to delete your local db to get a refresh one.

marcus-bcl: Cheers. That's actually taken care of by a change I made here: https://github.com/ministryofjustice/hmpps-delius-api/blob/bcaa9809daa9205a2a16c6c8023cfdfa01222927/src/main/resources/application-dev.yml#L12

The database file will automatically be cleaned/replaced when we change existing files.

marcus-bcl (approved):

PR #50, opened 2021-02-24
DAPi-87 - Allow non-selectable contact types with SPG override set to…

electrified:
… be used


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/entity/ContactType.kt
```diff
@@ -14,7 +14,7 @@ import javax.persistence.Table
 
 @Entity
 @Table(name = "R_CONTACT_TYPE")
-@Where(clause = "SELECTABLE = 'Y'")
+@Where(clause = "SELECTABLE = 'Y' OR SPG_OVERRIDE = 1")
```
mxco86, line 17: Should we be checking that it's not SELECTABLE when the override is set or does that not matter?

electrified: Checking the ND code, either selectable, override or both is acceptable.

mxco86 (approved):
LGTM

PR #53, opened 2021-02-25
feat: nsi business logic level validation

axle-h:


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/extensions/ProviderExtensions.kt
```diff
@@ -0,0 +1,16 @@
+package uk.gov.justice.digital.hmpps.deliusapi.service.extensions
+
+import uk.gov.justice.digital.hmpps.deliusapi.entity.Provider
+import uk.gov.justice.digital.hmpps.deliusapi.entity.Staff
+import uk.gov.justice.digital.hmpps.deliusapi.entity.Team
+import uk.gov.justice.digital.hmpps.deliusapi.exception.BadRequestException
+
+fun Provider.getTeamAndStaffOrBadRequest(teamCode: String, staffCode: String): Pair<Team, Staff> {
```
electrified, line 8: Again 2 operations combined into one method, could be split?

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/dto/v1/NewNsiManager.kt
```diff
@@ -6,26 +6,14 @@ import uk.gov.justice.digital.hmpps.deliusapi.validation.StaffCode
...
   @ApiModelProperty("The code of the team managing this intervention")
   @field:TeamCode
   val team: String,
 
-  /**
-   * TODO must be existing, selectable (via selectable flag) provider
-   */
   @ApiModelProperty("The code of the provider managing this intervention")
   @field:ProviderCode
   val provider: String,
```
electrified, line 19: The provider can be derived from the staffCode, so is this redundant?

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/extensions/NsiTypeExtensions.kt
```diff
@@ -0,0 +1,44 @@
...
+}
+
+private fun NewNsi.getLevel(): NsiTypeLevel {
+  if (this.eventId != null) {
+    return NsiTypeLevel.EVENT
+  }
+  return NsiTypeLevel.OFFENDER
+}
+
+fun NsiType.assertSupportedLevel(newNsi: NewNsi) {
```
electrified, line 32: Contact types also have the concept of event/ offender level, we should probably enforce on there too. I will add a story to the backlog...

electrified (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/dto/v1/NewNsiManager.kt
electrified: Just seen Andrews comment, so ignore this.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/dto/v1/NewNsiManager.kt
```diff
@@ -6,26 +6,14 @@ import uk.gov.justice.digital.hmpps.deliusapi.validation.StaffCode
 import uk.gov.justice.digital.hmpps.deliusapi.validation.TeamCode
 
 data class NewNsiManager(
```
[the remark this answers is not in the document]
axle-h: This just makes mapping easier & cleaner from an api perspective ie manager: { staff: "", provider: "", ... } instead of { managerStaff: "", managerProvider: "" ... }. Makes it cleaner for us to expose new fields later like for eg if a client needed a reference to a manger by id. You're right though this is all negligible - it was just more effort to abstract it further than the existing "one active manager per nsi" abstraction for slightly more boilerplate - which I can live with.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/extensions/OffenderExtensions.kt
```diff
@@ -0,0 +1,25 @@
+package uk.gov.justice.digital.hmpps.deliusapi.service.extensions
+
+import uk.gov.justice.digital.hmpps.deliusapi.entity.Event
+import uk.gov.justice.digital.hmpps.deliusapi.entity.Offender
+import uk.gov.justice.digital.hmpps.deliusapi.entity.Requirement
+import uk.gov.justice.digital.hmpps.deliusapi.exception.BadRequestException
+
+fun Offender.getEventAndRequirement(eventId: Long?, requirementId: Long?): Pair<Event?, Requirement?> {
```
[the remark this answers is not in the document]
axle-h: This feels like service level code to me not data access level code (it throws bad request exceptions for eg) hence it's in the service package. Just needed a quick & dirty way to share this across the services. I'm still new to Kotlin but I 'think' stuff like this is one of the design goals of extension functions i.e. better placement of logic that only touches the public contract.
Happy to split this though - it was implemented at this level as it's the largest unit shared between the services.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/extensions/NsiTypeExtensions.kt
axle-h: probably be good to share this logic via a HasEventId interface on NewNsi & NewContact and a HasOffenderEventLevel interface on NsiType & Contact type

PR #55, opened 2021-02-26
feat: nsi status contact

axle-h:


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/NsiService.kt
```diff
@@ -119,18 +123,49 @@ class NsiService(
...
+      nsiId = nsi.id,
+      eventId = nsi.event?.id,
+      providerId = manager.provider?.id!!,
+      teamId = manager.team?.id!!,
+      staffId = manager.staff?.id!!,
+      timestamp = nsi.statusDate,
+    )
+    this.contactService.createSystemContact(status)
+
+    val referral = status.copy(typeId = null, typeCode = "NREF")
```
electrified, line 152: It would be good to extract these string literals out and put them in constants with descriptive names, they could then be referenced from the tests too.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/repository/AuditedInteractionRepository.kt
```diff
@@ -7,5 +7,5 @@ import java.util.Date
 
 @Repository
 interface AuditedInteractionRepository : JpaRepository<AuditedInteraction, Date> {
-  fun findAllByUserId(userId: Long): List<AuditedInteraction>
+  fun findAllByUserIdAndBusinessInteractionCode(userId: Long, businessInteractionCode: String): List<AuditedInteraction>
```
electrified, line 10: This is OK as it's only used for the tests. Maybe worth a note here? Reads from the table are SUPER SLOW in the real envs and should be avoided.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/ContactService.kt
```diff
@@ -116,4 +120,53 @@ class ContactService(
     val entity = contactRepository.saveAndFlush(contact)
     return ContactMapper.INSTANCE.toDto(entity)
   }
+
+  /**
+   * Adds a system generated contact record.
+   * System generated contacts are non-audited & non-validated (other than referential integrity of course).
+   */
+  fun createSystemContact(request: NewSystemContact) {
```
electrified, line 128: It's a bit of a shame we have code duplication with the other create contact method.

electrified (approved):

electrified:
Looks good!

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/ContactService.kt
axle-h: Don't really see a nice way around it without over complicating both

PR #56, opened 2021-02-26
DAPI-55: H2 database updates to support NSI endpoint

MichaelWetherallBCL:


marcus-bcl (changes requested):
As discussed, will hold off on merging until the new outcome data is added

marcus-bcl (approved):

PR #59, opened 2021-03-01
fix: manager team & staff are optional

axle-h:


axle-h:
Also includes [DAPI-93](https://jira.engineering-dev.probation.hmpps.dsd.io/browse/DAPI-93)

src/test/kotlin/uk/gov/justice/digital/hmpps/deliusapi/integration/IntegrationTestBase.kt
```diff
@@ -74,24 +77,6 @@ abstract class IntegrationTestBase {
...
-  protected fun shouldAudit(interaction: AuditableInteraction, parameters: Map<String, Any?>) {
-    val interactions = auditedInteractionRepository.findAllByUserIdAndBusinessInteractionCode(userId, interaction.code)
-    val expected = parameters.entries.map { "${it.key}='${it.value}'" }.toSet()
-
-    assertThat(interactions).anyMatch {
-      it.success &&
-        it.businessInteraction.code == interaction.code &&
-        it.parameters.split(",").map { p -> p.trim() }.toSet().containsAll(expected)
-    }
-  }
```
axle-h, line 93: I tried a few things but ended up giving up with integration testing auditing.
Issue is that there's no canonical primary key on the table so just like Delius, we tagged the datetime up instead. Hibernate is doing weird things with this. It's either not writing audit records that fall on the same second or when selecting it is overwriting results from it's local cache. In the worst case we're replicating Delius behaviour so I think it's fine to just drop this for now. Maybe look at raw JDBC queries for auditing at some point. Yay.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/entity/ProviderUser.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import java.io.Serializable
+import javax.persistence.Column
+import javax.persistence.Embeddable
+import javax.persistence.EmbeddedId
+import javax.persistence.Entity
+import javax.persistence.Table
+
+@Embeddable
+data class ProviderUserIdentity(
```
electrified, line 11: I wasn't expecting any security changes in this PR based on the title! :) Which JIRA are these bits for?

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/contact/ContactService.kt
```diff
@@ -66,6 +68,8 @@ class ContactService(
     val provider = providerRepository.findByCode(request.provider)
       ?: throw BadRequestException("Provider with code '${request.provider}' does not exist")
 
+    securityUserContext.grantProvider(provider.id)
```
electrified, line 71: I feel like these could get out of control if we have to litter them all over the codebase.

See later comment re Spring Security annotations.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/security/SecurityUserContext.kt
```diff
@@ -19,4 +23,12 @@ class SecurityUserContext(
     val token = authentication
     return if (token.authSource == "delius") token.userId ?: defaultStaffId else defaultStaffId
   }
+
+  fun grantProvider(providerId: Long) {
```
electrified, line 27: This feels like we're implementing our own security system in parallel with spring security.

The check itself seems fine, but can we wrap it up in a way so we can just annotate service methods with appropriate spring security annotations?

https://docs.spring.io/spring-security/site/docs/5.2.x/reference/html/authorization.html#access-control-using-preauthorize-and-postauthorize

The question then becomes at what level to add the annotations.. Might have to do some restructuring?

electrified (changes requested):
The team and staff bits look OK, but less sure about the security parts and think we should try and leverage spring security more.

Possibly split into 2 PRs?

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/security/SecurityUserContext.kt
axle-h: Definitely agree, newish to Spring so missed this one. Problem here is that the provider is retrieved and saved in one method. Maybe be ok to repeat this though and the security check (however it's implemented) returns ok when the provider does not exist - the service method can then throw. The won't matter where it's annotated then - although maybe be best to be on the request DTO if possible. This will select the provider twice but doesn't seem like an issue,.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/entity/ProviderUser.kt
axle-h: I'll split this off

axle-h:
Removed commit containing security work, will reimplement with Spring security in a followup PR

electrified (approved):

PR #61, opened 2021-03-03
feat: validate user-provider access

axle-h:


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/entity/Provider.kt
```diff
@@ -10,7 +10,6 @@ import javax.persistence.Table
 
 @Entity
 @Table(name = "PROBATION_AREA")
-@Where(clause = "SELECTABLE = 'Y'")
```
axle-h, line 13: ProviderUser can be associated to a non-selectable provider :-(

electrified (approved):
Great work :)

PR #67, opened 2021-03-05
Initial work on DAPI-108 - using database user

electrified:


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/repository/UserRepository.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Query
+import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.deliusapi.entity.User
+import java.time.LocalDate
+
+@Repository
+interface UserRepository : JpaRepository<User, Long> {
+  @Query("select u from User u where u.distinguishedName = ?1 and ( ?2 < u.endDate or u.endDate = null)")
```
marcus-bcl, line 11: Distinguished name is case-insensitive. There is an index on `UPPER(DISTINGUISHED_NAME)` so you should be able to use that for that comparison.

mxco86 (approved):
👍

marcus-bcl (approved):

PR #73, opened 2021-03-12
feat: enforcement

axle-h:


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/contact/ContactService.kt
```diff
@@ -87,6 +90,15 @@ class ContactService(
     entity.description = request.description
     entity.notes = getNotes(entity.notes, request.notes)
 
+    if (request.enforcement != entity.enforcements.getOrNull(0)?.action?.code) {
```
electrified, line 93: Wondering about this check where it only gets the first enforcement from the collection - should it check if any of the enforcements match?

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/extensions/ContactExtensions.kt
```diff
@@ -0,0 +1,28 @@
...
+import java.time.Duration
+
+fun Contact.getDuration(): Duration {
+  if (startTime != null && endTime != null) {
+    return Duration.between(startTime, endTime)
+  }
+  return Duration.ZERO
+}
+
+fun Contact.updateOutcome(value: ContactOutcomeType?) {
```
electrified, line 16: Where is this called from?

src/test/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/contact/CreateContactTest.kt
```diff
@@ -43,6 +43,7 @@ class CreateContactTest : ContactServiceTestBase() {
       .hasProperty(Contact::offender, offender)
       .hasProperty(Contact::type, type)
       .hasProperty(Contact::outcome, outcome)
+      .hasProperty(Contact::enforcements, listOf(enforcement))
```
electrified, line 46: Add the other new contact fields into here too?

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/extensions/ContactExtensions.kt
```diff
@@ -0,0 +1,28 @@
...
+  outcome = value
+  attended = value?.attendance
+  complied = value?.compliantAcceptable
+
+  // If offender has complied and attended (outcome is acceptable) then set any hours credited
+  if (type.recordedHoursCredited && attended == true && complied == true) {
```
electrified, line 22: Feel like we should have an explicit test covering this logic?

src/main/resources/db/data/V1__reference_data/V1_1__base_reference_data.sql
```diff
@@ -37245,6 +37245,7 @@ INSERT INTO R_ENFORCEMENT_ACTION (ENFORCEMENT_ACTION_ID, CODE, DESCRIPTION, SELE
...
 INSERT INTO R_ENFORCEMENT_ACTION (ENFORCEMENT_ACTION_ID, CODE, DESCRIPTION, SELECTABLE, CONTACT_TYPE_ID, OUTSTANDING_CONTACT_ACTION, CREATED_BY_USER_ID, ROW_VERSION, LAST_UPDATED_USER_ID, CREATED_DATETIME, LAST_UPDATED_DATETIME, TRAINING_SESSION_ID, RESPONSE_BY_PERIOD, SPG_OVERRIDE) VALUES (2500001531, 'LC03', 'Alternative to Recall Letter Sent (Pilot Only)', 'Y', 2500034731, 'Y', 1, 3, 1, TO_DATE('2019-05-09 21:46:31', 'YYYY-MM-DD HH24:MI:SS'), TO_DATE('2019-05-10 23:25:47', 'YYYY-MM-DD HH24:MI:SS'), null, null, null);
+INSERT INTO R_ENFORCEMENT_ACTION (ENFORCEMENT_ACTION_ID, CODE, DESCRIPTION, SELECTABLE, CONTACT_TYPE_ID, OUTSTANDING_CONTACT_ACTION, CREATED_BY_USER_ID, ROW_VERSION, LAST_UPDATED_USER_ID, CREATED_DATETIME, LAST_UPDATED_DATETIME, TRAINING_SESSION_ID, RESPONSE_BY_PERIOD, SPG_OVERRIDE) VALUES (2500001532, 'ROM', 'Refer to Offender Manager', 'Y', 658, 'Y', 1, 7, 1, '2013-02-05 10:08:52', '2019-05-10 23:25:47', null, 7, 0);
```
electrified, line 37248: Do we enforce anywhere that only the ROM code is valid at the moment?

src/test/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/contact/CreateContactTest.kt
```diff
@@ -151,6 +152,14 @@ class CreateContactTest : ContactServiceTestBase() {
     shouldThrowBadRequest()
   }
 
+  @Test
```
electrified, line 155: There are a few success cases we don't have tests for now, e.g. a contact without an enforcement.

I think we should add tests to cover these other valid scenarios.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/contact/ContactService.kt
axle-h: Should have mentioned, the relationship is annotated one-to-one in delius, using same meta here doesnt seem to work for inserting it via a JpaRepository though. I think it's because the foreign key is not on the primary keys - ie it's only OneToOne at the application level. Pretty sure I tried every permutation of OneToOne annotation options and couldn't get it to work.
So hacking around with the collection as if it is OneToOne seems best approach. Do you suggest maybe throwing if we find > 1?

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/extensions/ContactExtensions.kt
axle-h: ah good spot

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/extensions/ContactExtensions.kt
axle-h: yeah might spin it out like the other stuff, difficult to test in the contact service

src/main/resources/db/data/V1__reference_data/V1_1__base_reference_data.sql
axle-h: Yeah the reference data will be restricted to ROM enforcement only so to replicate this here would be redundant (and annoying in the future!)

src/test/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/contact/CreateContactTest.kt
axle-h: Gone over it again and I think it's just the inverse of this that's not tested
```
    if (enforcement != null) {
      enforcement.contact = contact
      contact.enforcements.add(enforcement)
    }
```
Do you agree? I know this test class has shrunk massively but the fact it was so unreasonable was the reason for splitting off the validation service, which now captures all the fine grained data driven success & failures.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/contact/ContactService.kt
electrified: Interesting that similar annotations from delius don't work here!

Throwing an exception if there is more than one seems like a good safety net.

src/test/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/contact/CreateContactTest.kt
axle-h: And missing nsi, although not sure that's ever been tested...

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/contact/ContactService.kt
axle-h: Yeah they have it as ManyToOne on Enforcement side 

https://github.com/ministryofjustice/delius/blob/b57df1981aaee802dc951229f41e08c8aa11b52f/NDelius-lib/src/main/java/uk/co/bconline/ndelius/entity/sentencing/Enforcement.java#L70

And OneToOne on Contact side

https://github.com/ministryofjustice/delius/blob/b57df1981aaee802dc951229f41e08c8aa11b52f/NDelius-lib/src/main/java/uk/co/bconline/ndelius/entity/contactlog/Contact.java#L258

Which doesn't build in our version of spring data.

src/test/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/contact/CreateContactTest.kt
electrified: There is the "Creating nsi contact" test for testing with NSIs. We don't have a tests fro an unoutcomed contact, although we do test for that at the integration level

electrified (approved):

PR #75, opened 2021-03-16
feat: enforcement action system contact

axle-h:


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/entity/AuditedInteraction.kt
```diff
@@ -13,7 +13,7 @@ import javax.persistence.Table
 
 @Entity
 @Table(name = "AUDITED_INTERACTION")
-data class AuditedInteraction(
+class AuditedInteraction(
```
electrified, line 16: Reading up on it, I don't think this is a bad change. The implementation of equals/hashcode data classes provide aren't how you're supposed to implement them for JPA, so I don't think we've lost much tbh.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/entity/Contact.kt
```diff
@@ -128,15 +128,21 @@ data class Contact(
   @Column(name = "TRUST_PROVIDER_FLAG", nullable = false)
   var trustProviderFlag: Long = 0,
 
+  /**
+   * TODO set this from configuration
+   */
   @Column(name = "STAFF_EMPLOYEE_ID", nullable = false)
-  var staffEmployeeId: Long = 0,
+  var staffEmployeeId: Long = 1,
```
electrified, line 135: I'm going to create a ticket to mop up these hardcoded bits... think we should tackle it soon.

electrified (approved):

PR #84, opened 2021-03-24
DAPI-53 Implement transactions

electrified:


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/advice/AuditableAspect.kt
```diff
@@ -17,6 +19,7 @@ annotation class Auditable(val interaction: AuditableInteraction)
 
 @Aspect
 @Component
+@Order(Ordered.LOWEST_PRECEDENCE - 1)
```
electrified, line 22: By default the transaction runs at the lowest precedence.

Running at one higher means our audit stuff runs before the transaction on the way in, and after on the way out.

The could be any value lower than Integer.MAX_VALUE (Ordered.LOWEST_PRECEDENCE) and would work

marcus-bcl (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/advice/AuditableAspect.kt
marcus-bcl: Nice. Is it worth adding that as a comment in the code?

axle-h (approved):

PR #87, opened 2021-03-29
feat: update nsi

axle-h:


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/nsi/NsiSystemContactService.kt
```diff
@@ -0,0 +1,148 @@
...
+import java.time.LocalDateTime
+import java.time.LocalTime
+import javax.transaction.Transactional
+
+/**
+ * Convenience service for handling the (many) system contacts generated when maintaining NSIs.
+ * Note: All contacts created in this service should be rolled back if primary entity cannot be saved.
+ *       To enforce this, all operations require an existing transaction.
+ */
+@Transactional(Transactional.TxType.MANDATORY)
```
electrified, line 22: I switched all the transactional annotation to the Spring transaction annotation - the equivalent of this would be   `@Transactional(propagation = Propagation.MANDATORY)`

We should be consistent and use one or the other, the Spring one offers a few more settings - readonly, isolation levels and timeout, but otherwise it makes no difference.

axle-h: Pushed a fix, was in use in NsiManagerService also

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/nsi/NsiSystemContactService.kt
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  fun updateCommencedContact(nsi: Nsi, request: UpdateNsi) {
+    when {
+      nsi.startDate == null && request.startDate != null ->
+        // nsi is now commenced
+        createWellKnownContact(nsi, WellKnownContactType.NSI_COMMENCED, request.startDate)
+      nsi.startDate != null && request.startDate == null ->
+        // delete commencement contact
+        contactRepository.deleteByNsiIdAndTypeCode(nsi.id, WellKnownContactType.NSI_COMMENCED.code)
```
electrified, line 64: The delius delete contact operation does quite a few other things - deleting child contacts, recalculating FTC counts,  breach status. Can we be sure we don't need this logic too? I guess most of it doesn't apply to this contact type?

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/nsi/NsiSystemContactService.kt
```diff
@@ -0,0 +1,149 @@
...
+
+  fun createStatusContact(nsi: Nsi) = createStatusContact(nsi, nsi.status!!, nsi.statusDate)
+
+  fun updateStatusContact(nsi: Nsi, newStatus: NsiStatus, request: UpdateNsi) {
+    if (nsi.status?.id != newStatus.id) {
+      // create new status
+      createStatusContact(nsi, newStatus, request.statusDate)
+    } else if (nsi.statusDate != request.statusDate) {
+      // status unchanged but status date updated
+      val existing = contactRepository.findAllByNsiIdAndTypeId(nsi.id, newStatus.contactTypeId)
```
electrified, line 40: Should this have a date check? I.e. only update contacts matching the old date?

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/nsi/NsiSystemContactService.kt
MichaelWetherallBCL: It might not apply to that contact type for now but contact types are highly configurable and may need those bits later. Also it wont be wasted work as the delete contact function could be called by other processes.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/nsi/NsiSystemContactService.kt
axle-h: yeah, it should, otherwise it will update the date of all statuses..

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/nsi/NsiSystemContactService.kt
axle-h: there are no FTC, breach or child contacts for this commencement contact. If this is a future requirement then I'd say leave it for then. Otherwise we will have all the delete logic in the app but unused & untested.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/nsi/NsiSystemContactService.kt
axle-h: pushed a fix

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/nsi/NsiSystemContactService.kt
axle-h: pushed up something that should address the safety of this -> it will now delete all recursive linked contacts & throw if any deleted contact would have affected FTC.

electrified (approved):

PR #88, opened 2021-03-30
DAPI-154 - Add staff endpoint

electrified:


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/dto/v1/staff/NewStaff.kt
```diff
@@ -0,0 +1,25 @@
...
+  @field:Size(max = 35)
+  val forename: String,
+
+  @ApiModelProperty("The provider code of the staff member")
+  @field:NotBlank
+  @ProviderCode
+  val provider: String,
+
+  val teams: List<String>?,
+)
```
axle-h, line 25: These field names conflict with the NSI PR, I used more sensible names.
image
What do you think?

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/dto/v1/staff/StaffDto.kt
```diff
@@ -0,0 +1,13 @@
+package uk.gov.justice.digital.hmpps.deliusapi.dto.v1.staff
+
+import java.time.LocalDate
+
+data class StaffDto(
+  val code: String,
+  var startDate: LocalDate,
+  val surname: String,
+  val forename: String,
+  val private: Boolean,
```
axle-h, line 10: Does this not cause any issues?! Would prefer not to use reserved words where possible.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/dto/v1/staff/NewStaff.kt
electrified: Sure, happy to go with your naming

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/dto/v1/staff/StaffDto.kt
electrified: Haha, it works fine. IntelliJ's syntax highlighting is a bit more intelligent than githubs and doesn't colour it differently. I can change it though.

image

axle-h: Ah nice, nah if there's no issues using this name then I'm sure it'll be fine. I suppose we'd also want to ensure our clients can handle it ok too. Does Java in community-api have issues with a field named 'private' ?

axle-h: Tyepscript can handle it: https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgCIFcC2mCeyDeAUMicgA5TABuckAXMgMpiUgDmA3MaQPZgAW0ACr9QbBs1adCAX0KEEPEAGcwyAB4MM2PAF4C5SjXrIA5BWq0IpgDTI+gqCLENTD6AFoBY08hldFFR4AGwgAOmCeNgAKdQBKDiA

PR #92, opened 2021-04-07
poc e2e tests

axle-h:
A new e2e task has been added to the delius-api project. This will generate an open api document and then use this to generate an api client. The e2e project itself is:

* Implemented in JUnit5 & Kotlin.
* Uses the generated client.
* Can use application entities & repositories for database access in order to create test data.
* Can call stored procedures to clean up.

We should start by getting back to parity with the existing postman implemented tests.

### TODO

- [x] The incremental build is not working
- [x] Authentication (simple oauth2.0 client will do)
- [x] Sugar to streamline authentication, creating existing data, database access if required
- [x] Port all tests over to the new project
- [x] Setup codebuild to run


src/e2e/kotlin/uk/gov/justice/digital/hmpps/deliusapi/InstrumentationTest.kt
```diff
@@ -0,0 +1,26 @@
...
+
+  private val client = OperationHandlerApi(basePath)
+
+  @Test
+  fun `Getting health`() {
+    val response = client.getHealth(null) as Map<String, String>
+    assertThat(response)
+      .isNotNull
+      .containsEntry("status", "UP")
+  }
```
axle-h, line 17: had a go (and have a stash of the attempt) at getting the Health contract into the spec but was taking too long. Springfox do not make it easy if you don't have a controller that you can annotate.

build.gradle.kts
```diff
@@ -3,16 +3,95 @@ plugins {
...
+    dependsOn += generateApiSpec
+    inputs.file(deliusApiSpec)
+    outputs.dir("$projectDir/$generatedKotlinSrc")
+  }
+  val compileGeneratedKotlin by getting {
+    dependsOn += openApiGenerate
+  }
+  val e2e by getting {
+    dependsOn += compileGeneratedKotlin
+  }
```
axle-h, line 43: 1. **generateApiSpec** - spin up a mock api via JUnit & Spring test runner, get the spec and save to disc. This is madness but the process has been lifted from the [sprinfox examples](https://github.com/springfox/springfox-demos/tree/master/boot-static-docs).
2. **openApiGenerate** - generates a kotlin client from the spec into src/generated, see the config for this step below
3. ** compileGeneratedKotlin** - compile the generated client
4. **e2e** use the generated client to call & test the api

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/config/SwaggerConfig.kt
```diff
@@ -49,3 +57,50 @@ class SwaggerConfig {
...
+              .split("/")
+              .joinToString("") { it.capitalize() }
+        } else uniqueId
+        Optional.of(result)
+      }
+      .get()
+
+    builder.uniqueId(uniqueId)
+    builder.codegenMethodNameStem(uniqueId)
+  }
```
axle-h, line 88: this is just renaming the endpoints
* The health & info endpoints are just called handle & handle1
* All the others have {javaMethodName}Using{HttpMethod} & this must be unique, I've renamed all the controller methods to be unique & verbose and stripped off the Using{HttpMethod} bit

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/config/SwaggerConfig.kt
```diff
@@ -1,21 +1,32 @@
...
 import springfox.documentation.spring.web.plugins.Docket
+import springfox.documentation.swagger.common.SwaggerPluginSupport
 import java.time.LocalTime
 
 @Configuration
 class SwaggerConfig {
   @Bean
   fun deliusApi(buildProperties: BuildProperties): Docket =
-    Docket(DocumentationType.SWAGGER_2)
+    Docket(DocumentationType.OAS_30)
```
axle-h, line 29: is it an issue if we only support OAPI 3.0?

electrified: I don't think so. Matt has put OpenAPI3 in the ADR.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/config/SwaggerConfig.kt
```diff
@@ -49,3 +79,46 @@ class SwaggerConfig {
       emptyList()
     )
 }
+
+@Component
+@Order(SwaggerPluginSupport.SWAGGER_PLUGIN_ORDER + 1000) // we need this to run after annotation & default readers
+class NiceOperationReader : OperationBuilderPlugin {
```
electrified, line 85: Wondering if there is a more descriptive name than "nice"? Somthing about improving the naming, or making it human readable...

electrified (approved):

PR #94, opened 2021-04-09
DAPI-157 - return conflict for appointment clashes

electrified:


src/test/kotlin/uk/gov/justice/digital/hmpps/deliusapi/integration/v1/contact/CreateContactTest.kt
```diff
@@ -41,32 +42,32 @@ class CreateContactTest : IntegrationTestBase() {
...
-          of(valid.copy(offenderCrn = "1234567"), "offenderCrn"),
-          of(valid.copy(type = ""), "type"),
-          of(valid.copy(type = "12345678910"), "type"),
-          of(valid.copy(provider = "1234"), "provider"),
-          of(valid.copy(team = "1234567"), "team"),
-          of(valid.copy(staff = "12345678"), "staff"),
-          of(valid.copy(officeLocation = "12345678"), "officeLocation"),
-          of(valid.copy(requirementId = 1L, eventId = null), "requirementId cannot be provided without also providing eventId"),
-          of(valid.copy(requirementId = 2500083652, nsiId = 2500018597), "nsiId cannot be provided when requirementId is also provided"),
+          of(valid.copy(offenderCrn = ""), "offenderCrn", HttpStatus.BAD_REQUEST),
```
electrified, line 45: This isn't ideal... having a default for the parameter, or making it nullable and only supplying it when not bad request didn't work.

Open to suggestions!

axle-h: one thing Ive done in other complex tests is to wrap up all arguments in a TestCase data class. You can then easily have default values for stuff.

axle-h: I dont think this is actually 'that' bad though

axle-h (approved):

PR #95, opened 2021-04-12
Document e2e tests & re-enable patch tests

axle-h:


bin/end-to-end-test
```diff
@@ -5,12 +5,6 @@ OAUTH2_SECRET=
...
 
-docker run -v "$(pwd)":/etc/newman \
-	--network="host" \
-	-t postman/newman run 'Delius API.postman_collection.json' \
-	--env-var "oauth2Client=$OAUTH2_CLIENT" \
-	--env-var "oauth2Secret=$OAUTH2_SECRET"
-
 docker run -v "$(pwd)":/etc/newman \
 	--network="host" \
 	-t postman/newman run 'Community API.postman_collection.json' \
```
axle-h, line 10: why do we have CAPI e2e tests in this repository?

electrified (approved):

bin/end-to-end-test
electrified: I think it was just for convenience - so you could see a full end to end from community API, as Matt was trying to provide it out for Interventions usage

PR #98, opened 2021-04-13
delete contact

axle-h:


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/advice/AuditableAspect.kt
```diff
@@ -15,7 +15,7 @@ import uk.gov.justice.digital.hmpps.deliusapi.service.audit.AuditableInteraction
 
 @Target(AnnotationTarget.FUNCTION)
 @Retention(AnnotationRetention.RUNTIME)
-annotation class Auditable(val interaction: AuditableInteraction)
+annotation class Auditable(vararg val interactions: AuditableInteraction)
```
axle-h, line 18: the delete operation audits a different interaction type based on contact date

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/security/ProviderResponseAuthority.kt
```diff
@@ -1,8 +0,0 @@
-package uk.gov.justice.digital.hmpps.deliusapi.security
-
-import org.springframework.security.access.prepost.PostAuthorize
-import uk.gov.justice.digital.hmpps.deliusapi.config.Authorities
-
-@Retention(AnnotationRetention.RUNTIME)
-@PostAuthorize("hasAuthority('${Authorities.PROVIDER}'.concat(returnObject.provider))")
-annotation class ProviderResponseAuthority
```
axle-h, line 8: This didnt actually work.. The failure was masked by the fact that the only place it was used was in a private function, which is ignored by AOP security.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/security/SecurityUserContext.kt
```diff
@@ -13,4 +15,11 @@ class SecurityUserContext {
     }
 
   fun getCurrentDeliusUserId() = authentication.userId
+
+  fun assertProviderAuthority(provider: String) {
+    val providerAuthority = Authorities.PROVIDER + provider
+    if (!authentication.authorities.any { it.authority == providerAuthority }) {
+      throw AccessDeniedException("No access to provider with code '$provider'")
+    }
+  }
```
axle-h, line 24: open to better solutions for implementing this here. Cause of this is that I need a way to run security within a method where breaking out a service to trigger AOP security doesn't make sense.

electrified (approved):

PR #99, opened 2021-04-14
separate e2e envs & enable database assert on test env

axle-h:


src/e2e/kotlin/uk/gov/justice/digital/hmpps/deliusapi/config/EndToEndTestActiveProfilesResolver.kt
```diff
@@ -0,0 +1,11 @@
+package uk.gov.justice.digital.hmpps.deliusapi.config
+
+import org.springframework.test.context.ActiveProfilesResolver
+
+class EndToEndTestActiveProfilesResolver : ActiveProfilesResolver {
+  override fun resolve(testClass: Class<*>): Array<String> {
+    val env = System.getenv("SPRING_PROFILES_ACTIVE")?.split(',', ' ')?.toTypedArray()
```
marcus-bcl, line 7: Could this be achieved by adding `spring.profiles.include: e2e` to the config files?

marcus-bcl (approved):

src/e2e/kotlin/uk/gov/justice/digital/hmpps/deliusapi/config/EndToEndTestActiveProfilesResolver.kt
electrified: I wondered about this too - is it so you can optionally run against different DBs by setting additional profiles using the env var, without modifying any of the committed files?

src/e2e/kotlin/uk/gov/justice/digital/hmpps/deliusapi/v1/contact/CreateContactV1Test.kt
```diff
@@ -23,13 +25,15 @@ class CreateContactV1Test @Autowired constructor(
   private lateinit var created: Contact
 
   @Test
+  @Transactional
```
electrified, line 28: Did the test not work without the transaction annotation?

For mine I think I will have to leave this off, as test transactions roll back, which will revert the deletion.

electrified (approved):

src/e2e/kotlin/uk/gov/justice/digital/hmpps/deliusapi/config/EndToEndTestActiveProfilesResolver.kt
axle-h: afraid not, here's the default: https://github.com/spring-projects/spring-framework/blob/master/spring-test/src/main/java/org/springframework/test/context/support/DefaultActiveProfilesResolver.java
Makes sense to me for unit & integration tests TBF as you would have all the environment configu hard coded.

src/e2e/kotlin/uk/gov/justice/digital/hmpps/deliusapi/v1/contact/CreateContactV1Test.kt
axle-h: :-(
This is required to have a hibernate session present for lazy properties to work. There may be an alternative to this, like wrapping the database assert bit only in an explicit transaction.

src/e2e/kotlin/uk/gov/justice/digital/hmpps/deliusapi/config/EndToEndTestActiveProfilesResolver.kt
axle-h: @electrified yeah, this is meant for selecting which environment to use, which all have different reference data. This reminds me that the buildspec.yml should be updated to activate the `test` profile.

axle-h: Just pushed a fix. btw there is documentation for this in: `doc/testing.md`

PR #101, opened 2021-04-14
DAPI-158 - Add tests for staff and team, generate team codes

electrified:


src/e2e/resources/application-local-h2.yml
```diff
@@ -63,3 +63,18 @@ e2e:
       subType: KSS026
       status: REFER
       length: 1
+  staffs:
```
electrified, line 66: Not keen on this incorrect plural but it was the easiest option :D

src/e2e/kotlin/uk/gov/justice/digital/hmpps/deliusapi/config/ConfigurationExtensions.kt
```diff
@@ -60,3 +62,29 @@ fun EndToEndTestConfiguration.newNsi(select: NsiSelector): NewNsi {
...
+  val team = select(teams)
+  return NewTeam(
+    cluster = team.cluster,
+    description = team.description,
+    ldu = team.ldu,
+    provider = team.provider,
+    type = team.type,
+    unpaidWorkTeam = team.unpaidWorkTeam,
+  )
+}
```
axle-h, line 90: Do we get anything for having environment specific configuration on staff first & last names, team description?
If we do then it's a shame having to have this over using the generated request DTOs just to satisfy Kotlin being closed by default...

axle-h (approved):

PR #103, opened 2021-04-16
init breach & ftc e2e tests

axle-h:


src/e2e/resources/application-local-h2.yml
```diff
@@ -1,80 +0,0 @@
-e2e:
-  url: "http://localhost:8080"
-  databaseAssert: false # disabled as we expect api+h2 to be running in a container TODO see what we can do
-  oauth:
-    url: "http://localhost:9090/auth"
```
axle-h, line 5: Was getting annoying having to maintain these & the lack of reliable db access meant some of the more involved tests were pointless. The integration tests also already covered all test cases & more.
Only value I can see with them is to support developers without access to the oracle image for whatever reason.. we have this file in source control now if we want it back at some point.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/repository/ContactRepository.kt
```diff
@@ -54,6 +54,12 @@ interface ContactRepository : JpaRepository<Contact, Long> {
   fun findAllByNsiIdAndTypeIdAndDate(nsiId: Long, typeId: Long, date: LocalDate): List<Contact>
 
   fun findAllByNsiIdAndTypeCode(nsiId: Long, typeCode: String): List<Contact>
+
+  fun deleteAllByEventIdAndTypeNationalStandardsContactIsTrue(eventId: Long)
+
+  fun deleteAllByEventIdAndTypeCode(eventId: Long, typeCode: String)
```
axle-h, line 60: these are dangerous but only ever used by the e2e tests... be good to have a way of disabling their implementation by default but couldnt see anything

electrified (approved):

PR #114, opened 2021-05-24
DST-8977 Deploy to Pre-Prod and Perf environments

marcus-bcl:


.circleci/config.yml
```diff
@@ -193,11 +193,27 @@ workflows:
             - hmpps-common-vars
             - hmpps-delius-deploy-to-ecs-stage
           requires:
-            - request-stage-approval
+            - request-pre-prod-approval
```
electrified, line 196: It deploys to stage and preprod at the same time?

marcus-bcl: Yes. Same as we do for community-api: https://github.com/ministryofjustice/community-api/blob/main/.circleci/config.yml#L141

electrified (approved):

PR #119, opened 2021-06-03
DAPI-186 set rar requirement flag

axle-h:


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/contact/ContactValidationService.kt
```diff
@@ -227,4 +228,31 @@ class ContactValidationService(
...
   }
+
+  /**
+   * Validates the RAR requirement flag on the request.
+   * The RAR flag is required when:
+   * `contact linked to non-terminated requirement with type category "F" (either directly or through nsi)`
+   *   AND (
+   *     `contact type has rar requirement flag set`
+   *     OR `contact is linked to nsi, which is linked to non-terminated rar requirement` AND `contact type has attendance flag set`
+   *   )
```
mahalrmoj, line 239: Is this checked below "which is linked to non-terminated rar requirement"?

axle-h: yeah, it's a bit convoluted, note that `val linkedRequirement = requirement ?: nsi?.requirement`
the terminated status is checked `!linkedRequirement.isTerminated(request.date)`
and to determine whether the requirement came from the nsi & is attendance contact: `nsi != null && type.attendanceContact`

mahalrmoj (approved):

PR #120, opened 2021-06-08
Calculate RAR counts

electrified:


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/contact/ContactValidationService.kt
```diff
@@ -253,6 +253,6 @@ class ContactValidationService(
   }
 
   companion object {
-    val RAR_CODE = "F"
+    const val RAR_CODE = "F"
```
axle-h, line 256: can this not be used in the repository queries?

PR #123, opened 2021-06-10
WIP investigating opening classes and hibernate lazy loading

electrified:


marcus-bcl:
Closing as we're not looking into this at the moment.  Will leave the branch intact in case we need to investigate Hibernate lazy-loading performance issues again in future.

PR #136, opened 2021-08-13
DAPI-221 Rescheduling a rescheduled appointment fails

electrified:
Example scenario:
1. Create appointment running from 10:00 - 11:00
2. Reschedule this appointment to be from 10:30 - 11:30
	-> This rescheduled appointment is in a clashed state with the
	original appointment, but we ignore it during this creation as
	we have the original apppointment id and exclude it from the
	clash check
3. Attempt to reschedule the rescheduled appointment to be 14:00 - 15:00
	-> This fails when we try and set an outcome on the appointment
	we are attempting to reschedule (the one created in step 2). It clashes
	with the original appointment as we are not excluding that original
	appointment ID as we did when it was created. We don't have the context of what
	that original appointment ID was either

Solution: When updating an appointment, don't perform the clash check if
the time/ date isn't changing. The apppointment might already be in a
clashed state, which isn't our fault.


axle-h (approved):

mahalrmoj (approved):
LGTM.

PR #142, opened 2021-09-03
Moved OracleDB docker scripts to separate repo

marcus-bcl:
New location: https://github.com/ministryofjustice/hmpps-delius-docker-images/tree/main/oracledb-18c

There is also a WebLogic image available in the new repository, to enable devs to run the Delius UI as well as the DB: https://github.com/ministryofjustice/hmpps-delius-docker-images/tree/main/weblogic-14c


mxco86 (approved):
LGTM

PR #161, opened 2021-11-01
DAPI-281 - Escape special characters in free text search query

joseph-bcl:
Fixes elastic search errors caused by searching with special characters such as colon or hyphen.


marcus-bcl:
Is it possible to write an integration test for this?  It runs against a real Elasticsearch via TestContainers, so I guess escaping the query should work in the tests too?

marcus-bcl (approved):

PR #164, opened 2021-11-05
Add ADR proposal for URL structure rules

mxco86:
This is a proposal for the URL structure ADR. Happy to revise based on the discussion here


doc/architecture/decisions/0010-ensure-consistent-urls.md
```diff
@@ -0,0 +1,60 @@
...
+
+Delius API URLs must adhere to the following rules:
+
+1. The URL must use the version number of the endpoint in the form
+   `/v{version_number}` as the first path element 
+2. All resources in URLs must be the pluralised version of the resource name
+3. Where the resource name is an acronym the URL reference should be the
+   acronym with an appended 's' character 
+4. Where a request input requires a CRN the URL must contain the path
+   `/offender/{crn}` directly after the version path part
```
marcus-bcl, line 29: Should this be `offenders` plural?
```suggestion
   `/offenders/{crn}` directly after the version path part
```

marcus-bcl:
How do we feel about endpoints with actions in the URL e.g. contact search is currently `/v1/contact/search` - what could we change that to if we want to stay resource-based?

I can't think of anything other than `/v1/offenders/{crn}/contacts/search`...

electrified:
> How do we feel about endpoints with actions in the URL e.g. contact search is currently `/v1/contact/search` - what could we change that to if we want to stay resource-based?
> 
> I can't think of anything other than `/v1/offenders/{crn}/contacts/search`...

If search returned a list of the same entities we use on the other endpoints, I'd say we could make it `GET /v1/offenders/{crn}/contacts?search=blah` or something, but as search has it's own `SearchContact` response, that might make it confusing that it's not a list of `ContactDto`s

mxco86:
Agree that we should use GET and query parameters for search and there aren't that many parameters so I don't think this would be a problem. Also agree that the response means it should probably move to it's own resource (maybe something like `/v1/offenders/crn/contactsearch?q=blah`)?

marcus-bcl:
We followed the approach for probation-offender-search and prisoner-offender-search of using POST with parameters in the body.  I think the reasoning behind it is that the query may contain sensitive information, and while query parameters are encrypted over TLS, they are usually included in access logs etc which might not be treated sensitively. I'm not sure if that's a good enough reason not to use query parameters though, as like you say the URLs make more sense using query parameters.

mxco86:
OK yes I see the reasoning but I'm not sure it's completely valid - more obscurity than safety. Using URL parameters does mean we have to be a bit more careful with logging and the automated telemetry collection but I don't think these are logged by default to e.g. App Insights so you would have to make an effort to include them. The concern then is the clients but no client should be logging free-form user input as this is really a serious bug.

doc/architecture/decisions/0010-ensure-consistent-urls.md
mxco86: Good spot!

marcus-bcl (approved):

PR #165, opened 2021-11-08
Bugfix/dapi 282

robertmccormackbconline:


marcus-bcl (approved):
Looks good to me.

@axle-h @electrified does this look okay to you?  It's just moving the `@Where` clause off the entity and into extension methods, so we can fetch inactive child entities when we need to.  It has the side effect of moving the filtering into the application layer, but I don't think that should be a problem?

PR #166, opened 2021-11-10
DAPI-129 ObjectOptimisticLockingFailureException results in 409 Conflict rather than 500 Internal Server Error

mahalrmoj:
**What does this pull request do?**
When multiple "notification contact" post requests are made concurrently to delius-api for the same NSI, one of the requests will fail and Delius API will return a status that indicates there has been a data conflict (409).

**What is the intent behind these changes?**
The response allows a consumer (such as community-api) to distinguish between this scenario and other failures scenarios and act accordingly.

**Any breaking changes?**
None. However in the case of an optimistic locking failure a 409 is returned instead of an internal server error


mxco86 (approved):
👍

marcus-bcl (approved):

PR #169, opened 2021-11-24
DAPI-278

joseph-bcl:
- Added validation to ensure Enforcement Actions can only be added if they are linked to the Contact Type Outcome


mxco86 (approved):
LGTM

PR #178, opened 2021-12-16
DAPI-311 : Fixed 500 response when contact outcome type is inactive

robertmccormackbconline:


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/entity/ContactOutcomeType.kt
```diff
@@ -36,4 +34,8 @@ class ContactOutcomeType(
   @Column(name = "ENFORCEABLE", length = 1)
   @Type(type = "yes_no")
   var enforceable: Boolean?,
+
+  @Column(name = "SELECTABLE", nullable = false)
+  @Type(type = "yes_no")
+  var selectable: Boolean,
```
marcus-bcl, line 40: Is there a constraint for this in the DB?  i.e. are we certain it will never be null?

src/test/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/contact/CreateContactTest.kt
```diff
@@ -262,18 +270,22 @@ class CreateContactTest : ContactServiceTestBase() {
...
-      false -> outcomeMock.thenThrow(BadRequestException("bad outcome"))
+    if (havingValidActiveContactOutcomeType) {
+      when (havingValidOutcomeType) {
+        null -> outcomeMock.thenReturn(null)
+        true -> outcomeMock.thenReturn(outcome)
+        false -> outcomeMock.thenThrow(BadRequestException("bad outcome"))
+      }
+    } else {
+      outcomeMock.thenCallRealMethod()
     }
```
marcus-bcl, line 289: Why do we need to call the real method for an inactive outcome?  Should we not just mock another BadRequestException?

e.g.
```suggestion
    if (havingValidActiveContactOutcomeType) {
      when (havingValidOutcomeType) {
        null -> outcomeMock.thenReturn(null)
        true -> outcomeMock.thenReturn(outcome)
        false -> outcomeMock.thenThrow(BadRequestException("bad outcome"))
      }
    } else {
      outcomeMock.thenThrow(BadRequestException("inactive outcome"))
    }
```

Which could maybe be tidied up a bit, e.g. renaming the var to just `havingActiveOutcomeType`:
```suggestion
    when (havingValidOutcomeType) {
      null -> outcomeMock.thenReturn(null)
      true -> when (havingActiveOutcomeType) {
         true -> outcomeMock.thenReturn(outcome)
         false ->  outcomeMock.thenThrow(BadRequestException("inactive outcome"))
      }
      false -> outcomeMock.thenThrow(BadRequestException("bad outcome"))
    }
```

I guess the same applies to the recent office location changes too?

src/test/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/contact/UpdateContactTest.kt
```diff
@@ -268,9 +278,13 @@ class UpdateContactTest : ContactServiceTestBase() {
...
 
-  private fun havingOutcomeType(having: Boolean = true) {
+  private fun havingOutcomeType(having: Boolean = true, isActive: Boolean = true,) {
     val mock = whenever(validationService.validateOutcomeType(request, contact.type))
-    if (having) mock.thenReturn(outcome) else mock.thenThrow(BadRequestException("bad outcome"))
+    if (isActive) {
+      if (having) mock.thenReturn(outcome) else mock.thenThrow(BadRequestException("bad outcome"))
+    } else {
+      mock.thenCallRealMethod()
+    }
```
marcus-bcl, line 287: Same here, would it be better to mock an exception?
```suggestion
    if (!having) {
      mock.thenThrow(BadRequestException("bad outcome"))
    } else if (!isActive) {
      mock.thenThrow(BadRequestException("inactive outcome"))
    } else {
      mock.thenReturn(outcome)
    }
```

src/test/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/contact/ContactValidationServiceTest.kt
```diff
@@ -89,6 +97,14 @@ class ContactValidationServiceTest {
   fun `Successfully validating required-or-non-required office location without office location`() =
     attemptingToValidateOfficeLocation(success = null, locationFlag = YesNoBoth.B, havingRequestOfficeLocation = false)
 
+  @Test
+  fun `Successfully validating active office location`() =
+    attemptingToValidateOutcomeType(success = true, isActive = true)
```
marcus-bcl, line 102: Should this be:
```suggestion
    attemptingToValidateOfficeLocation(success = true, isActive = true)
```

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/entity/ContactOutcomeType.kt
robertmccormackbconline: Yes, there is a constraint in the DB on selectable for contact_outcome_type, the constraint name varies in each env (eg: SYS_C0023214 on ST9) but it definitely is there

src/test/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/contact/ContactValidationServiceTest.kt
```diff
@@ -554,13 +572,15 @@ class ContactValidationServiceTest {
...
     havingOfficeLocation: Boolean = true,
+    isActive: Boolean = true,
   ) {
     val request = Fake.newContact().copy(
       officeLocation = if (havingRequestOfficeLocation) "some-office-location" else null
     )
     val type = Fake.contactType().apply { this.locationFlag = locationFlag }
     val officeLocation = Fake.officeLocation().apply {
       code = if (havingOfficeLocation) "some-office-location" else "some-other-office-location"
+      endDate = if (isActive) Fake.randomFutureLocalDate().atStartOfDay() else Fake.randomLocalDateTime()
```
marcus-bcl, line 583: If it's inactive, do we need to ensure the endDate is set to a date in the past?  Is there a `Fake.randomPastLocalDate()`?

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/entity/ContactOutcomeType.kt
marcus-bcl: Nice - thanks

src/test/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/contact/ContactValidationServiceTest.kt
robertmccormackbconline: There is, however Fake.randomPastLocalDate() returns a LocalDate object whereas Fake.randomLocalDateTime returns the same date but in LocalDateTime format which is what we need, might be worth renaming the function in this case?

 `fun randomPastLocalDate(): LocalDate = faker.date().past(10, 1, TimeUnit.DAYS).toLocalDate()`
 `fun randomLocalDateTime(): LocalDateTime = faker.date().past(10, 1, TimeUnit.DAYS).toLocalDateTime()`
  
  [https://github.com/ministryofjustice/hmpps-delius-api/blob/66541df9d3f3d6f8adff1189bed64a7acf30d0c9/src/test/kotlin/uk/gov/justice/digital/hmpps/deliusapi/util/Fake.kt#L100-L101](url)

marcus-bcl (approved):

PR #179, opened 2021-12-17
Updated to Java 17 & Kotlin 1.6, switched out Springfox for Springdocs

axle-h:


Dockerfile
```diff
@@ -7,7 +7,7 @@ WORKDIR /app
 ADD . .
 RUN ./gradlew assemble -Dorg.gradle.daemon=false
 
-FROM openjdk:11-jre-slim
+FROM openjdk:17-slim
```
marcus-bcl, line 10: Is there a reason for dropping `jre` here?   I had to add that in to get past a Trivy vulnerability in libgmp.

axle-h: There isn't a 17-jre-slim tag available for the openjdk image. Community API is using this 17-slim image and trivy is currently passing.

marcus-bcl (approved):

PR #180, opened 2021-12-17
DAPI-296 - Preserve Search terms in audit

joseph-bcl:
When using Contact Search an Audit record is created containing the parameters of the search.


marcus-bcl (approved):
@axle-h are you happy with this approach?  Feel like it could get unwieldy as we add more varied audit parameters for each new interaction... maybe an AuditContext per interaction would be better? Or just a Map...?

axle-h (approved):

axle-h:
> 

Agreed this is getting unwieldy, let's merge this now and address later?

PR #182, opened 2021-12-24
DAPI-314 Resolve various vulnerabilities

marcus-bcl:


axle-h (approved):

build.gradle.kts
```diff
@@ -138,7 +139,7 @@ dependencies {
   implementation("org.flywaydb:flyway-core:7.15.0")
   implementation("com.zaxxer:HikariCP:5.0.0")
   implementation("com.h2database:h2")
-  implementation("com.oracle.database.jdbc:ojdbc10:19.12.0.0")
+  implementation("com.oracle.database.jdbc:ojdbc11")
```
tomsmyers, line 142: should this be pinned to a specific version?

tomsmyers (approved):

build.gradle.kts
marcus-bcl: It's covered by Spring dependency management: https://docs.spring.io/spring-boot/docs/current/reference/html/dependency-versions.html#:~:text=ojdbc11-,21.3.0.0,-com.oracle.database

marcus-bcl: Had a bit more of a think about this... we use Oracle Database 19c so you're right maybe we should be pinning to the 19.x driver versions.  However Oracle recommends always using the same version or higher, and states driver 21.x is backward-compatible with 19c, so we should be fine to stick with the managed version: https://www.oracle.com/uk/database/technologies/faq-jdbc.html#:~:text=What%20is%20the%20JDBC%20and%20RDBMS%20interoperability%20matrix%20or%20the%20certification%20matrix%3F

PR #188, opened 2022-01-12
PI-12 Fix end-to-end tests for VPD

marcus-bcl:


src/e2e/kotlin/uk/gov/justice/digital/hmpps/deliusapi/EndToEndTest.kt
```diff
@@ -85,31 +86,33 @@ abstract class EndToEndTest {
   }
 
   private fun getToken(): String {
-    if (CACHED_TOKEN?.expired() == false) {
-      return CACHED_TOKEN!!.accessToken
+    if (CACHED_TOKEN?.expired() != false) {
```
mxco86, line 89: Can you remove the condition completely here?

doc/testing.md
```diff
@@ -59,18 +59,36 @@ database for test setup & teardown. Environment configuration is handled by
...
 
-To run the end-to-end tests against the local docker oracle database, use the
-`local-oracle` spring profile.
-This is configured by default to use the `delius-api-client` client from the auth service.
+To run the end-to-end tests against local Oracle database and Elasticsearch 
+containers, use the `local-oracle` and `local-elasticsearch` Spring profiles.
 
 ```bash
-SPRING_PROFILES_ACTIVE=local-oracle ./gradlew e2e
+SPRING_PROFILES_ACTIVE=local-oracle,local-elasticsearch ./gradlew e2e
```
mxco86, line 68: These profiles being slightly different from the application is a bit confusing - might be worth picking one or the other

mxco86 (approved):
LGTM - couple of small things

src/e2e/kotlin/uk/gov/justice/digital/hmpps/deliusapi/EndToEndTest.kt
marcus-bcl: I don't think so, it's effectively saying - if the CACHED_TOKEN is `null`, or CACHED_TOKEN.expired() is `true`, then fetch the token.  It is confusing though, maybe being more explicit is better e.g.
```suggestion
    if (CACHED_TOKEN == null || CACHED_TOKEN.expired()) {
```

doc/testing.md
marcus-bcl: Yeah agreed, it's not super obvious.  Maybe we just have `application-local.yml` for running against local containers, and `application-test.yml` for running against AWS Test?

src/e2e/kotlin/uk/gov/justice/digital/hmpps/deliusapi/EndToEndTest.kt
mxco86: OK yes - that is clearer

doc/testing.md
mxco86: Yep makes sense - we can do another ticket for rationalising these

mxco86 (approved):

PR #198, opened 2022-01-19
Fix appointment clashes in contact search e2e tests

marcus-bcl:
Updates the contact search test to use a simple type, rather than an appointment type, to prevent intermittent test failures if clashing contacts are added manually to the test environment.


trozzomoj (approved):

mxco86 (approved):
👍

PR #201, opened 2022-01-25
DAPI-309 + DAPI-310 Add Audit to Add Recall and Add Document Endpoints

joseph-bcl:
- Added auditing to the Add Document and Add Recall Endpoints
- Changed the AuditService to map to AuditContext instead of AuditParameters so that endpoints can use AuditContext subclasses that only contain relevant AuditParameters to that interaction
-  Added 'Add Recall' BI to V1_3__business_interactions.sql
- Updated Tests


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/audit/AuditContext.kt
```diff
@@ -1,27 +1,114 @@
...
+
+data class ProviderAuditContext(
+  var providerId: Long? = null
+) : AuditContext() {
+  override fun toMap() = mapOf(PROVIDER_ID to providerId)
+}
+
+data class DocumentAuditContext(
+  var alfrescoDocumentId: String? = null,
+  var contactId: Long? = null
```
marcus-bcl, line 89: Should this be a part of a `ContactDocumentAuditContext` class?  We're adding a new endpoint soon for uploading documents to court reports, so I expect we'd need to support different identifier here (i.e. `courtReportId`). Maybe it would be better to have generic `entityId` and `tableName` fields instead of `contactId`?

However, looking at Delius code - it only seems to add the `IWPDocumentId` parameter.  So maybe if we went with `documentId` instead of `contactId` and logged that as `IWPDocumentId` - so it's consistent?  We could still add in the `alfrescoDocumentId` after a successful upload, but it'd be useful to keep the existing IDs as they are.

marcus-bcl (approved):

PR #212, opened 2022-02-15
Licence Conditions fixes for Add Recall Endpoint

jamesmacbeth-unilink:
DAPI-332 - Added accepted/rejected decision and rejection reason to terminated licence condition transfers. This is to stop the transfers from still appearing on the Consolidated Transfer Review screen in NDelius.
DAPI-334 - Added condition to only delete licence condition-linked future contacts that do not already have an outcome.
DAPI-335 - Insert System Generated Contact when licence condition transfer is terminated.


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/sentencing/LicenceConditionService.kt
```diff
@@ -84,11 +90,76 @@ class LicenceConditionService(
...
+  }
+
+  private fun createSystemGeneratedContactForTerminatedTransfer(transfer: LicenceConditionTransfer) {
+    val licenceCondition = transfer.licenceCondition
+
+    val notes = listOfNotNull(
+      "Transfer Status: ${transfer.transferStatus.description}",
+      "Transfer Reason: ${transfer.allocationReason?.description}",
+      "Rejection Reason: ${transfer.rejectionReason?.description}",
+      "Owning Provider: ${transfer.originTeam?.provider?.description}",
```
marcus-bcl, line 146: What does it output when these are null?  Should it default to "N/A" or something, what does Delius do?

marcus-bcl (approved):
Approved with a question about defaults in the notes field.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/sentencing/LicenceConditionService.kt
jamesmacbeth-unilink: At the moment it will output "null" and Delius would do the same.
These values shouldn't be null (even though they're nullable in the database) as they're mandatory fields in ND and/or set in the previous function in DAPI.

PR #216, opened 2022-02-17
DAPI-318

madan-thapa:
DAPI-318 Its possible to perform a DAPI Return to Custody when the case is Post Sentence Supervision (ND doesnt allow this)


marcus-bcl:
@madan-thapa @jamesmacbeth-unilink is there an overlap between DAPI-318 and DAPI-327 (https://github.com/ministryofjustice/hmpps-delius-api/pull/214) - are these PRs doing the same thing or are they both required?

jamesmacbeth-unilink:
> @madan-thapa @jamesmacbeth-unilink is there an overlap between DAPI-318 and DAPI-327 (#214) - are these PRs doing the same thing or are they both required?

Yeah they both appear to be doing the same thing. I think what's happened here is the story DAPI-327 was raised in response to the defect DAPI-318 and both were scheduled separately. 
They both appear to be doing the same thing, albeit in slightly different ways, so I think either could be closed.

marcus-bcl:
> > @madan-thapa @jamesmacbeth-unilink is there an overlap between DAPI-318 and DAPI-327 (#214) - are these PRs doing the same thing or are they both required?
> 
> Yeah they both appear to be doing the same thing. I think what's happened here is the story DAPI-327 was raised in response to the defect DAPI-318 and both were scheduled separately. They both appear to be doing the same thing, albeit in slightly different ways, so I think either could be closed.

Makes sense, thanks @jamesmacbeth-unilink.  I guess this is the easier one to close, as the other has another change in it?  If so, could you close this one?

jamesmacbeth-unilink:
Closing as a duplicate of #214.

PR #217, opened 2022-02-17
Dapi 300 add document to equality

peter-bcl:


peter-bcl:
Created in new branch to remove vpd

PR #218, opened 2022-02-18
DAPI-300-EqualityDoc

peter-bcl:


marcus-bcl (commented):
Looks good to me.  Although @trozzomoj do you think it needs to be rewritten using the approach you used for court report documents?

mxco86:
Closing this one until we have a client use case. We'll keep the branch as a reference

PR #227, opened 2022-03-01
DAPI-301 - Insert Disability

mlaskowski4:


src/test/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/person/DisabilityServiceTest.kt
```diff
@@ -0,0 +1,164 @@
...
+      it.crn = crn
+    }
+    whenever(offenderRepository.findByCrn(crn))
+      .thenReturn(if (offenderExists) offender else null)
+  }
+
+  private fun havingDisabilityTypeCode(
+    code: String,
+    havingReferenceCode: Boolean,
+    selectable: Boolean
```
marcus-bcl, line 127: There's quite a bit of duplication above of `havingReferenceCode = true, selectable = true`.  Would the code read better if you added some defaults here?
```suggestion
    code: String = request.disabilityTypeCode,
    havingReferenceCode: Boolean = true,
    selectable: Boolean = true
```

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/dto/v1/person/NewDisability.kt
```diff
@@ -0,0 +1,23 @@
+package uk.gov.justice.digital.hmpps.deliusapi.dto.v1.person
+
+import io.swagger.v3.oas.annotations.media.Schema
+import uk.gov.justice.digital.hmpps.deliusapi.validation.DeliusDate
+import java.time.LocalDate
+import javax.validation.constraints.Size
+
+class NewDisability(
+  @Schema(title = "The date of the recall")
+  var disabilityTypeCode: String,
```
marcus-bcl, line 10: `The date of the recall` looks like it could be a copy-paste error?

Also, could we rename this field to just `NewDisability.type`?  It would be more consistent with other places (e.g. `NewContact.type`) - you can mention in the `@Schema` annotation that we expect a code value if needed.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/dto/v1/person/DisabilityDto.kt
```diff
@@ -0,0 +1,12 @@
+package uk.gov.justice.digital.hmpps.deliusapi.dto.v1.person
+
+import java.time.LocalDate
+
+data class DisabilityDto(
+  val id: Long,
+  val crn: String,
+  val disabilityType: String,
```
marcus-bcl, line 8: Similarly to the `NewDisability.disabilityTypeCode` comment - could this be renamed to `DisabilityDto.type`?

marcus-bcl (commented):
Couple of small comments, otherwise looks good.

Could do with finding out who is going to be using these endpoints though, and why.  So we can ensure it meets their needs before merging.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/person/DisabilityService.kt
```diff
@@ -0,0 +1,62 @@
...
+  @Auditable(AuditableInteraction.ADD_DISABILITY)
+  fun createDisability(crn: String, request: NewDisability): DisabilityDto {
+    val offender = offenderRepository.findByCrnOrNotFound(crn)
+
+    val audit = AuditContext.get(AuditableInteraction.ADD_DISABILITY) as OffenderAuditContext
+    audit.offenderId = offender.id
+
+    val disabilityType = referenceDataMasterRepository.findByCodeOrThrow(WellKnownReferenceDataset.DISABILITY_TYPE.code)
+      .standardReferences?.find { it.code == request.type && it.selectable }
+      ?: throw BadRequestException("An invalid Disability Type was provided.")
```
marcus-bcl, line 41: Could you swap this for a call to the `ReferenceDataService` ? It was added recently to simplify ref data lookups.

Should end up looking something like this:
```suggestion
    val disabilityType = referenceDataService.getReferenceData(WellKnownReferenceDataset.DISABILITY_TYPE, selectableOnly = true)
```

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/referencedata/WellKnownDisabilityType.kt
```diff
@@ -0,0 +1,5 @@
+package uk.gov.justice.digital.hmpps.deliusapi.service.referencedata
+
+enum class WellKnownDisabilityType(val code: String) {
```
marcus-bcl, line 3: Do we need this class (and the corresponding `getReferenceData` method) ?  It doesn't look like it's being used anywhere.

mlaskowski4: The class and corresponding method removed

marcus-bcl:
Thanks @mlaskowski4, the changes look good.  The equalities/disabilities endpoints are intended for use by the Assess Risks & Needs service, however we've been told they won't be using it at the moment.  So until they're ready we're going to hold off on merging this.

mxco86:
Closing this as we don't have a client use case. We will keep the branch as a reference

PR #230, opened 2022-03-03
PI-91 Refactor reference data lookups

marcus-bcl:
Previously, we were doing this every time we needed a single item of standard reference data:
```
    val allocationReason = referenceDataMasterRepository
      .findByCodeOrThrow(WellKnownReferenceDataset.NSI_MANAGER_ALLOCATION_REASON.code)
      .standardReferences?.find { it.code == "IN1" && it.selectable }
      ?: throw RuntimeException("Cannot find NSI allocation reason with code 'IN1'")
```
i.e. fetching all reference data, then filtering for the code we wanted.

This change adds a `StandardReferenceRepository` to allow us to look up specific reference data items without pulling back the whole list. It also adds a `ReferenceDataService` to make it easier to lookup well-known reference data items, without having to specify the list they're in.


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/referencedata/ReferenceDataService.kt
```diff
@@ -0,0 +1,39 @@
...
+
+  private fun getWellKnownReferenceData(itemCode: String, listCode: String, selectableOnly: Boolean = false): StandardReference {
+    try {
+      return getReferenceData(itemCode, listCode, selectableOnly)
+    } catch (e: MissingReferenceDataException) {
+      throw MissingWellKnownReferenceDataException(e)
+    }
+  }
+
+  fun getReferenceData(item: WellKnownAcceptedDecision, selectableOnly: Boolean = false) = getWellKnownReferenceData(item.code, WellKnownReferenceDataset.ACCEPTED_DECISION.code)
```
trozzomoj, line 31: Looks all good to me. I m just not 100% convinced by the kotlin way of creating an overloading. I m more for specialised functions than same name functions with different type , order , etc.

trozzomoj (approved):

mxco86 (approved):
Looks good. I do think it's worth the complexity of the overloading to give us clearly named and grouped reference data in the code rather than strings etc.

PR #236, opened 2022-03-10
Dapi 331

robertmccormackbconline:


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/throughcare/RecallService.kt
```diff
@@ -187,6 +187,22 @@ class RecallService(
       previousPOM.active = false
       previousPOM.endDate = LocalDateTime.now()
       prisonOffenderManagerRepository.save(previousPOM)
+
+      var contactNotes = "\nTransfer Reason: ${allocationReason.description}\nTransfer Date: ${previousPOM.allocationDate}" +
+        "\nFrom Establishment: ${previousPOM.provider.description}\nFrom Team: ${previousPOM.team.description}" +
+        "\nFrom Officer: ${previousPOM.staff.getDisplayName()}\n"
```
marcus-bcl, line 193: Would it be cleaner to use a multi-line string here?
```suggestion
      var contactNotes = """
        Transfer Reason: ${allocationReason.description}
        Transfer Date: ${previousPOM.allocationDate}
        From Establishment: ${previousPOM.provider.description}
        From Team: ${previousPOM.team.description}
        From Officer: ${previousPOM.staff.getDisplayName()}
      """.trimIndent()
```

robertmccormackbconline: Wasn't aware it could be done like that, thanks.

PR #237, opened 2022-03-14
DAPI-348 New Notes aren't appended to ERCL Contact

Seanvalmonte:
New Notes arent appended to ERCL Contact


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/throughcare/RecallService.kt
```diff
@@ -151,7 +151,7 @@ class RecallService(
   private fun insertRecallContact(offenderId: Long, eventId: Long, request: NewRecall) {
     val reason = recallReasonRepository.findByCodeAndSelectableTrue(request.recallReason)
       ?: throw IllegalArgumentException("Recall Reason with code '${request.recallReason}' does not exist")
-    val newNotes = listOfNotNull(request.notes, "Reason for Recall: ${reason.description}").joinToString("\n")
+    val newNotes = listOfNotNull("Reason for Recall: ${reason.description}", "Notes: ${request.notes}", "-------------------------------").joinToString("\n")
```
marcus-bcl, line 154: None of these strings can be null, so `listOfNotNull` seems redundant here?  We're also no longer handling the case where `request.notes` is null. 

Looking at the Delius code it doesn't print the "Notes: ..." line at all if notes are null or an empty string, so I think you need something more along the lines of:
```suggestion
    val newNotes = listOfNotNull(
      "Reason for Recall: ${reason.description}",
      request.notes?.takeIf{ it.isNotEmpty() }?.let{ "Notes: $it" },
      "-------------------------------"
    ).joinToString("\n")
```
(there might be a cleaner way of writing this, but you get the gist 😄)

marcus-bcl (changes requested):

marcus-bcl (approved):

PR #275, opened 2022-04-12
Define delius database and API for developer portal

sldblog:
Adds a Backstage entity metadata file to this repository so that the
component can be added to the developer portal
(https://developer-portal.apps.live.cloud-platform.service.justice.gov.uk/catalog)

The portal crawls through all MoJ repos, looking for
`catalog-info.yaml`.

The intent is for repositories to self-describe dependencies so that

- discovering integrations
- discovering existing solutions

becomes much simpler. Strategically, this file will replace most of
https://github.com/ministryofjustice/hmpps-architecture-as-code/blob/main/src/main/kotlin/uk/gov/justice/hmpps/architecture/model/Delius.kt


catalog-info.yaml
```diff
@@ -0,0 +1,63 @@
...
+  tags:
+    - kotlin
+    - spring-boot
+spec:
+  type: service
+  lifecycle: production
+  owner: group:delius-api
+  system: system:hmpps-delius # system defined in hmpps-delius-api
+  providesApis:
+    - api:hmpps-delius
```
sldblog, line 42: Is this naming/reference good? There's a chance to define APIs semantically here -- eg. API for contacts, API for something else...

sldblog, line 42: Is this naming/reference good? There's a chance to define APIs semantically here -- eg. API for contacts, API for something else...

mxco86 (approved):
LGTM

sldblog:
Thanks @mxco86! I'm not authorised to merge, could you merge if it's all right, please?

catalog-info.yaml
mxco86: Yes I think it makes sense to start like this and split them out semantically as we develop things

PR #300, opened 2022-05-06
DAPI-348

Seanvalmonte:
DAPI Recall: Update Return to Custody for the DAPI created Recall, new notes arent appended to ERCL contact (and different placement of the Reason for Recall in the notes)


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/entity/Contact.kt
```diff
@@ -96,7 +95,6 @@ class Contact(
   var alert: Boolean? = false,
 
   @Column(name = "CREATED_DATETIME", nullable = false)
-  @CreatedDate
```
marcus-bcl, line 99: Could we add a comment somewhere stating why we had to remove this?  Just to ensure someone doesn't add it back in at some point without realising Delius relies on it to infer the contact/recall linkage.  Might even be worth including a link to the Delius code (https://github.com/ministryofjustice/delius/blob/2e43abfe3110801bd1c3093bcde5fa001eae38e6/NDelius-ejb/src/main/java/uk/co/bconline/ndelius/service/throughcare/ThroughcareServiceBean.java#L1230-L1232)

marcus-bcl (commented):
Looks good, thanks Sean.  It's not ideal that we have to do this, but our hands are tied by the Delius code...

Would it be possible to add a test that verifies the recall and the contact are created with the same date/time?  Happy to help with this if you need.

marcus-bcl (approved):

jamesmacbeth-unilink (approved):

PR #306, opened 2022-05-17
Pi 170

stevomcallister:


marcus-bcl (commented):
Couple of small comments but looks good.

Also can we remove the `OffenderManagerMapper` and `OffenderManagerDto` classes now?

marcus-bcl:
@anthony-britton-moj I think there may be some linting issues but for some reason the CI didn't run... could you try logging in to https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-delius-api , to make sure it's authorised?

For future reference you can fix any formatting by running `./gradlew ktlintFormat` - I'd recommend adding that as a pre-commit hook too, using `./gradlew ktlintApplyToIdea addKtlintFormatGitPreCommitHook`.

src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/controller/v1/AllocationController.kt
```diff
@@ -13,20 +13,24 @@ import org.springframework.web.bind.annotation.RequestMapping
...
 class AllocationController(
-  private val service: AllocationService,
+    private val service: AllocatePersonService,
+    private val allocateEventService: AllocateEventService,
+    private val allocateRequirementService: AllocateRequirementService,
 ) {
 
-  @PostMapping(consumes = [MediaType.APPLICATION_JSON_VALUE])
-  @Operation(description = "Creates a new allocation record")
+  @PostMapping("/allocation", consumes = [MediaType.APPLICATION_JSON_VALUE])
```
marcus-bcl, line 32: Sorry just spotted - this should be `/allocations` plural.

PR #339, opened 2022-06-15
DST-12003 - Fix constraint violation when updating Responsible Officer

jamesmacbeth-unilink:
Changed RO update to saveAndFlush to resolve unique constraint issue when creating a new RO on recall.


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/throughcare/RecallService.kt
```diff
@@ -258,7 +258,7 @@ class RecallService(
     // End date previous responsible officer
     if (currentRO != null) {
       currentRO.endDate = LocalDateTime.now()
-      responsibleOfficerRepository.save(currentRO)
+      responsibleOfficerRepository.saveAndFlush(currentRO)
```
marcus-bcl, line 261: @jamesmacbeth-unilink could you add a comment about why this is required?

marcus-bcl (changes requested):

marcus-bcl (approved):

PR #417, opened 2022-10-10
fix for legacy psr documents

stevomcallister:


src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/court/report/NewTechHelper.kt
```diff
@@ -23,8 +23,7 @@ class NewTechHelper(
     cipher.init(Cipher.ENCRYPT_MODE, secret)
 
     val encrypted = cipher.doFinal(input.toByteArray())
-    val base64Encoded = Base64.getEncoder().encode(encrypted)
-
-    return String(base64Encoded)
+    val base64Encoded = Base64.getEncoder().encodeToString(encrypted)
+    return URLEncoder.encode(base64Encoded, Charsets.UTF_8)
```
marcus-bcl, line 27: Does this work all the time?  I think I tried it before, but it still failed for certain documents. See https://github.com/ministryofjustice/hmpps-delius-api/blob/c4d54f01a898fda79d551f5c258f05a58b01de87/src/main/kotlin/uk/gov/justice/digital/hmpps/deliusapi/service/court/report/NewTechHelper.kt#L28

stevomcallister: no it didn't work. but then neither does the original code. Looks like it's being double URL encoded trying different things to fix it.

PR #426, opened 2022-10-14
Add trivy suppression for https://avd.aquasec.com/nvd/cve-2022-42889.

izzyL-DJ:


izzyL-DJ:
Needed as apache commons-text 1.9 bundled with application insights agent jar and therefore can't be upgraded at present.

